PR #89, opened 2022-02-15
[#64] WorkflowDispatch Inputs

jmfayard:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

### DIFF
--- a/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/WorkflowDispatch.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/WorkflowDispatch.kt
@@ -5,12 +5,11 @@ class WorkflowDispatch(
     val inputs: Map<String, Input> = emptyMap(),
 ) : Trigger() {
 
-    @Suppress("EnumNaming", "EnumEntryName")
     enum class Type {
-        choice,
-        environment,
-        boolean,
-        string,
+        Choice,
+        Environment,
+        Boolean,
+        String,
     }
 
     class Input(
--- a/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/WorkflowDispatch.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/WorkflowDispatch.kt
@@ -1,3 +1,23 @@
 package it.krzeminski.githubactions.domain.triggers
 
-class WorkflowDispatch : Trigger()
+// https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs
+class WorkflowDispatch(
+    val inputs: Map<String, Input> = emptyMap(),
+) : Trigger() {
+
+    @Suppress("EnumNaming", "EnumEntryName")
+    enum class Type {
+        choice,
+        environment,
+        boolean,
+        string,
+    }
+
+    class Input(
+        val description: String,
+        val required: Boolean,
+        val type: Type,
+        val options: List<String> = emptyList(),
+        val default: String? = null,
+    )
+}
--- a/library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
@@ -75,9 +75,9 @@ private fun StringBuilder.printIfHasElements(
     space: String = "  ",
 ) {
     if (!items.isNullOrEmpty()) {
-        appendLine("$space$name:")
+        appendLine("$name:".prependIndent(space))
         items.forEach {
-            appendLine("$space  - '$it'")
+            appendLine("  - '$it'".prependIndent(space))
         }
     }
 }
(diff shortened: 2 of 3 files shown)


library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/WorkflowDispatch.kt
```diff
@@ -1,3 +1,23 @@
 package it.krzeminski.githubactions.domain.triggers
 
-class WorkflowDispatch : Trigger()
+// https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs
+class WorkflowDispatch(
```
krzema12, line 4: Side-note: I was thinking about something more fancy in case of these inputs: on workflow builder level, calling a function like `val myInput = workflowDispatchInput("inputKey",...)` and then you could use it in string interpolation like `My input value is {{ $myInput }}` - it would render as `My input value is {{ github.event.inputs.inputKey }}`.

But your solution is a good one as a first step - it allows users to use this GH feature, and we can always incrementally add the API I mentioned. It needs some more prototyping to ensure it works as expected.

library/src/test/kotlin/it/krzeminski/githubactions/yaml/TriggersToYamlTest.kt
```diff
@@ -39,6 +39,60 @@ class TriggersToYamlTest : DescribeSpec({
...
+            // given
+            val triggers = listOf(
+                WorkflowDispatch(
+                    inputs = mapOf(
+                        "logLevel" to WorkflowDispatch.Input(
+                            description = "Log level",
+                            type = WorkflowDispatch.Type.choice,
+                            required = true,
+                            default = "warning",
+                            options = "info warning debug".split(" "),
```
krzema12, line 54:
```suggestion
                            options = listOf("info", "warning", "debug"),
```

krzema12 (changes requested):

library/src/test/kotlin/it/krzeminski/githubactions/yaml/TriggersToYamlTest.kt
```diff
@@ -39,6 +39,60 @@ class TriggersToYamlTest : DescribeSpec({
                 |workflow_dispatch:
             """.trimMargin()
         }
+
+        it("renders with all parameters") {
+
+            // given
+            val triggers = listOf(
+                WorkflowDispatch(
+                    inputs = mapOf(
```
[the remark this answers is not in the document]
jmfayard: I added an input with `type = String`.
What's an implicit feedback?
Having a default value of type = String?

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
```diff
@@ -40,11 +41,43 @@ private fun Schedule.toYaml() = buildString {
...
+    val space = "      "
+    appendLine("${space}description: '$description'")
+    appendLine("${space}type: ${type.toYaml()}")
+    appendLine("${space}required: $required")
+    if (default != null) appendLine("${space}default: '$default'")
+    printIfHasElements(options, "options", space = "      ")
+}.removeSuffix("\n")
+
+// WDT = WorkflowDispatch.Type
+private fun WDT.toYaml(): String = when (this) {
```
krzema12, line 65: This `WDT` looks a bit awkward :see_no_evil: Could we use a regular import of `WorkflowDispatch.Type` here, for the sake of being explicit and self-documenting?

krzema12: To make the code a bit shorter, I usually import `WorkflowDispatch.Type` so that its items are available without using `WorkflowDispatch.Type`. WDYT?

jmfayard: The issue was that an enum was `String` which clashes with `kotlin.String`.
I am now using Type.String instead

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
```diff
@@ -40,11 +40,35 @@ private fun Schedule.toYaml() = buildString {
...
+        appendLine("  inputs:")
+        for ((key, input) in inputs) {
+            appendLine("    $key:")
+            appendLine(input.toYaml())
+        }
+    }
+}.removeSuffix("\n")
+
+private fun WorkflowDispatch.Input.toYaml() = buildString {
+    val space = "      "
```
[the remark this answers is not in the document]
krzema12: @jmfayard just this one left.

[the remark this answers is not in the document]
krzema12: Ah, I see you used `prependIndent` but it's not what I had in mind :sweat_smile: I wanted to avoid having `${space}` in front. I'll prepare a snippet in a spare time.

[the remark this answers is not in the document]
krzema12: Yeah, it's hard to deal with all these indents... You know, let's go forward with the code as is, and #80 covers some cleanup around serializing to YAML. We have good test coverage so I'm not worried about regressions.

krzema12 (approved):

PR #92, opened 2022-02-16
generation: deprecated properties

jmfayard:
Deprecated properties based on the field `input.deprecationMessage`
https://github.com/actions/setup-node/blob/f099707f6e7a91b89cdebdeca599fc76cc8bf088/action.yml#L33


krzema12 (approved):
Thanks! :rocket:

PR #105, opened 2022-02-27
write multiple workflows from single script

NikkyAI:


library/src/main/kotlin/it/krzeminski/githubactions/yaml/WriteWorkflows.kt
```diff
@@ -0,0 +1,26 @@
+package it.krzeminski.githubactions.yaml
+
+import it.krzeminski.githubactions.domain.Workflow
+import kotlin.io.path.writeText
+
+fun writeWorkflows(
```
krzema12, line 6: I have troubles understanding the use case for this function. There's already `Workflow.writeToFile()`. If you want to write more than one, you can just iterate over them. `args` lets you just write a single workflow, but you can just choose the workflow you want. Is your goal having a single script (entry point) for generating all or chosen workflow? If yes, why can't the script write all workflows every time?

NikkyAI: The function writes all of them at once to file when invoked without args, the goal is to also have working consistency checks for all the generated workflows

NikkyAI: Using args is mainly for the consistency checks.. The user does not have to redirect the output into file anymore.. We can add more sanity checks.. Eg. Check that working for is in the root of the repo.

NikkyAI: If nothing else, i hope that the utilities i wrote for the tests helps in improving the rest of the code :P

krzema12: Ok, I get your point. The problem here is that nearly everyone using this library has its ideas on how to write the workflows to files :sweat_smile: I'm a bit reluctant to adding support for all of them in the library (already added one in https://github.com/krzema12/github-actions-kotlin-dsl/commit/10ed9642c362beeb73b612d7d99bed430dd4b998) because the public APIs cost, they add complexity and need to be tested and maintained. Instead I aim at providing some basic APIs, and you can then have a function like the one here to implement your approach (from what I see, your function uses library's already available public API right?). Once the library has some more users, I'm eager to analyze how they do the workflows writing, and then maybe some popular option will emerge for which it makes sense to add utilities/support in the library itself.

As a low-hanging fruit, what we could do that could help you is adding `addConsistencyCheck: Boolean` to
https://github.com/krzema12/github-actions-kotlin-dsl/blob/51a9780e1f4e265755a456a87b9718d74802f7d2/library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt#L56
which would simplify your function a bit. It's a cheap change within already existing API.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
```diff
@@ -16,8 +17,8 @@ fun Workflow.toYaml(addConsistencyCheck: Boolean = true): String {
             run("Install Kotlin", "sudo snap install --classic kotlin")
             run(
                 "Consistency check",
-                "diff -u '${targetFile.pathString.replace('\\', '/')}' " +
```
krzema12, line 19: Wow, I didn't know about `invariantSeparatorsPathString`, thanks! :) I'll make this change alone.

NikkyAI:
My function needs the consistency check on he yaml to pass some kind of id/name 
The rest can be done in the scripts if we need to, or i can be in a extension library maybe

krzema12:
> My function needs the consistency check on he yaml to pass some kind of id/name if that could be modified or be passed on as a lambda.. The rest can be done in the scripts if we need to, or it can be in a extension library maybe

This id/name would be there to run generation for a chosen workflow. What if your function would regenerate all workflows every time? What problem would it create? I think the performance is not an issue given that writing a single workflow is not computationally or IO expensive.

NikkyAI:
looking at how the selfcheck works..
```bash
"diff -u '.github/workflows/someworkflow.yaml' <('someScript.main.kts')
```

it relies on the stout output of the script to diff against

i just tried to stay compatible how (presumably) people use the library so far.. (because thats how the README explains the setup)


we could copy the original file, regenerate and compare them too
but that might cause weird issues or github being unhappy about the workflow modifying itself ?

krzema12:
Sorry, let's go back to my questions, otherwise I don't understand the original problem.

To visualize it, the proposed approach is to have:

```kotlin
            writeWorkflows(
                addConsistencyCheck = true,
                workflows = listOf(
                    workflow1,
                    workflow2
                ),
                "Test workflow 1",
            )
```

I propose writing
```kotlin
listOf(
    workflow1,
    workflow2,
).forEach {
    it.writeToFile(addConsistencyCheck = true)
}
```

(we'd need to add `addConsistencyCheck` to `writeToFile`, but we can do it). **Would that work for you?**

NikkyAI:
without generating individual to stdout how do we do consistency checks ?

if the yaml can still do a diff of itself against some output or generated file from the script then that works

jmfayard:
If the parameter`fun Workflow.writeToFile(addConsistencyCheck: Boolean = false)` can be added, that's all I ned.
The code says it can't be done currently, and I don't know what would need to be changed for that.

https://github.com/krzema12/github-actions-kotlin-dsl/blob/b8f252e4816b85cdb29c70a6561660b8c7f3d567/library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt#L56-L62

NikkyAI:
we could replace the check with a step running the script and then
```bash
git diff .github/workflows/workflowname.yaml | wc -l
```
as a empty output is 0 lines == no error

otherwise this works too

```bash
test $(git diff .github/workflows/workflowname.yaml | wc -l) -eq 0
```

krzema12:
@NikkyAI I like this approach! We could add `addConsistencyCheck` argument to `writeToFile` that would use this approach. About support for other platforms, I don't think we have to bother since we can run the consistency check job on UbuntuLatest.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
```diff
@@ -1,62 +1,62 @@
...
+        val consistencyCheckJob = this.toBuilder().job(
+            name = "check_yaml_consistency",
+            runsOn = UbuntuLatest,
+        ) {
+            uses("Check out", CheckoutV2())
+            run("Install Kotlin", "sudo snap install --classic kotlin")
+            run("Execute script", sourceFile.invariantSeparatorsPathString)
+            run(
+                "Consistency check",
+                "test \$(git diff '${targetFile.invariantSeparatorsPathString}' | wc -l) -eq 0"
```
krzema12, line 21: We want to support the previous consistency check logic in `toYaml`. The `git diff`-based one should be used only if using `writeToFile(true)`. Otherwise it's a breaking change and we force everyone to use the new method.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
```diff
@@ -1,62 +1,62 @@
...
+
+fun Workflow.toYaml(addConsistencyCheck: Boolean = true): String {
+    val jobsWithConsistencyCheck = if (addConsistencyCheck) {
+        val consistencyCheckJob = this.toBuilder().job(
+            name = "check_yaml_consistency",
+            runsOn = UbuntuLatest,
+        ) {
+            uses("Check out", CheckoutV2())
+            run("Install Kotlin", "sudo snap install --classic kotlin")
+            run("Execute script", sourceFile.invariantSeparatorsPathString)
```
krzema12, line 18: Would be good to remove the target file first. Otherwise there's a risk someone didn't actually call `writeToFile()` in the script, the script won't modify the target file so `git diff` will produce empty diff, and the consistency check will succeed where it shouldn't.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/WriteWorkflows.kt
```diff
@@ -0,0 +1,26 @@
+package it.krzeminski.githubactions.yaml
+
+import it.krzeminski.githubactions.domain.Workflow
+import kotlin.io.path.writeText
+
+fun writeWorkflowsOrPrintYaml(
```
krzema12, line 6: I thought we agreed to not add this function. Is it some leftover?

README.md
```diff
@@ -81,10 +81,10 @@ names with your own.
 
    Alternatively, apart from `toYaml()` which returns a string, there's also `writeToFile()` which puts the string
    straight into the file specified in workflow's `targetFile`. It may come in handy when operating on multiple
-   workflows and having automation that regenerates them. Note that this function doesn't provide consistency check.
-4. Generate the YAML by calling the above script and redirecting its output to the desired YAML file path:
```
krzema12, line 85: This guide should differentiate between both ways, e.g. removing `> .github/workflows/hello_world_workflow.yml` in case of using the stdout redirection is incorrect.

library/src/test/kotlin/it/krzeminski/githubactions/EndToEndTest.kt
```diff
@@ -1,494 +1,599 @@
...
+
+                run(
+                    name = "Hello world!",
+                    command = "echo 'hello!'",
+                )
+            }
+        }
+
+        // when
+        workflowWithTempTargetFile.writeToFile(addConsistencyCheck = false)
```
krzema12, line 273: Please also add tests for `writeToFile(addConsistencyCheck = true)`.

krzema12 (changes requested):

library/src/test/kotlin/it/krzeminski/githubactions/EndToEndTest.kt
NikkyAI: added a test for it

README.md
NikkyAI: i reverted my changes to the gide, only adjusting the section about `writeToFIle()` a bit, removing the note about not providing consistency checks

library/src/main/kotlin/it/krzeminski/githubactions/yaml/WriteWorkflows.kt
NikkyAI: i removed it

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
NikkyAI: good idea, i will do that

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
NikkyAI: sensible, i have made the required changes

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
```diff
@@ -6,19 +6,30 @@ import it.krzeminski.githubactions.domain.Workflow
...
-                    "<('${sourceFile.invariantSeparatorsPathString}')"
-            )
+            if (useGitDiff) {
+                run(
+                    "Execute script",
+                    "rm '$targetFile' && " + sourceFile.invariantSeparatorsPathString
+                )
+                run(
+                    "Consistency check",
+                    "test \$(git diff '${targetFile.invariantSeparatorsPathString}' | wc -l) -eq 0"
```
krzema12, line 24: How about simplifying it to
```suggestion
                    "git diff --exit-code '${targetFile.invariantSeparatorsPathString}'"
```

https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code

krzema12 (changes requested):

library/src/test/kotlin/it/krzeminski/githubactions/FixNewlines.kt
```diff
@@ -0,0 +1,9 @@
+package it.krzeminski.githubactions
+
+fun String.fixNewlines(): String {
```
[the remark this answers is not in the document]
NikkyAI: would `unixNewlines()` work better?
the goal is to replace `/r/n` from windows files with `/n` so that we can compare the text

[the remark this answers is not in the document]
NikkyAI: no clue how it worked before.. when i tested this locally on windows it failed.. (since then i worked on linux), i will have to switch to windows again and see again

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
NikkyAI: i completely overlooked that, good find

library/src/main/kotlin/it/krzeminski/githubactions/yaml/ToYaml.kt
```diff
@@ -6,19 +6,30 @@ import it.krzeminski.githubactions.domain.Workflow
 import it.krzeminski.githubactions.dsl.toBuilder
 import kotlin.io.path.invariantSeparatorsPathString
 
-fun Workflow.toYaml(addConsistencyCheck: Boolean = true): String {
+fun Workflow.toYaml(addConsistencyCheck: Boolean = true, useGitDiff: Boolean = false): String {
```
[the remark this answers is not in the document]
NikkyAI: sensible, i did that

library/src/test/kotlin/it/krzeminski/githubactions/EndToEndTest.kt
```diff
@@ -1,494 +1,597 @@
...
+                steps:
+                  - id: step-0
+                    name: Check out
+                    uses: actions/checkout@v2
+                  - id: step-1
+                    name: Hello world!
+                    run: echo 'hello!'
+        """.trimIndent()
+
+        // when
```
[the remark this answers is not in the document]
NikkyAI: i made a new test for it

library/src/test/kotlin/it/krzeminski/githubactions/FixNewlines.kt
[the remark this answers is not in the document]
NikkyAI: well.. i dunno why but it works without just fine.. and since its not necessary.. removed it again

library/src/main/kotlin/it/krzeminski/githubactions/domain/Workflow.kt
```diff
@@ -1,13 +1,13 @@
...
-
-data class Workflow(
-    val name: String,
-    val on: List<Trigger>,
-    val env: LinkedHashMap<String, String>,
-    val sourceFile: Path,
-    val targetFile: Path,
-    val jobs: List<Job>,
-)
+package it.krzeminski.githubactions.domain
```
krzema12, line 1: Let's avoid making whitespace changes. Here I think line endings changed?

library/src/test/kotlin/it/krzeminski/githubactions/EndToEndTest.kt
```diff
@@ -1,494 +1,690 @@
...
+                    name: Some step consuming other step's output
+                    uses: actions/checkout@v2
+                    with:
+                      repository: ${'$'}{{ step-0 }}
+                      ref: ${'$'}{{ steps.step-0.outputs.commit_sha }}
+                      token: ${'$'}{{ steps.step-0.outputs.my-unsafe-output }}
+        """.trimIndent()
+    }
+
+    test("writeToFile() - write multiple workflows") {
```
krzema12, line 568: Hmm, just wondering: what added value does this test have? It just calls writeToFile twice.

krzema12 (changes requested):
Minor ones.

library/src/main/kotlin/it/krzeminski/githubactions/domain/Workflow.kt
krzema12: In The tests file as well.

library/src/test/kotlin/it/krzeminski/githubactions/EndToEndTest.kt
NikkyAI: honestly, it was a different function before.. might as well remove it

library/src/main/kotlin/it/krzeminski/githubactions/domain/Workflow.kt
krzema12: #115

NikkyAI: i might have had my git misconfigured, it had `core.autocrlf` enabled in linux.. which messed some stuff up
not sure if i fixed that before or after i cloned that..

NikkyAI: should be fixed again

krzema12:
Looks good code-wise! One last thing to ensure: did you test it end-to-end, running on GitHub?

krzema12:
> Looks good code-wise! One last thing to ensure: did you test it end-to-end, running on GitHub?

BTW, having true end-to-end tests is in scope of https://github.com/krzema12/github-actions-kotlin-dsl/issues/87. The current EndToEndTest.kt is a misnomer, it's rather about integration testing.

NikkyAI:
i did not go though the effort to setup a repo for my fork to do a proper end-to-end test
but i guess i could verify that it "breaks"  verifying a older scripts, i shall do that

NikkyAI:
testing around a bit the error modes i managed to reproduce are..

1. file missing

```
Run git diff --exit-code '.github/workflows/docker_workflow.yml'
fatal: ambiguous argument '.github/workflows/docker_workflow.yml': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Error: Process completed with exit code 128.
```

2. file content differs
```diff
git diff --exit-code '.github/workflows/docker_workflow.yml'
diff --git a/.github/workflows/docker_workflow.yml b/.github/workflows/docker_workflow.yml
index 10c2366..c6fbb0b 100644
--- a/.github/workflows/docker_workflow.yml
+++ b/.github/workflows/docker_workflow.yml
@@ -13,25 +13,8 @@ on:
       - 'main'
 
 jobs:
-  "check_yaml_consistency":
-    runs-on: "ubuntu-latest"
-    steps:
-      - id: step-0
-        name: Check out
-        uses: actions/checkout@v2
-      - id: step-1
-        name: Install Kotlin
-        run: sudo snap install --classic kotlin
-      - id: step-2
-        name: execute script
-        run: rm '.github/workflows/docker_workflow.yml' && '.github/workflows/docker_workflow.main.kts'
-      - id: step-3
-        name: Consistency check
-        run: git diff --exit-code '.github/workflows/docker_workflow.yml'
   "build_job":
     runs-on: "ubuntu-latest"
-    needs:
-      - "check_yaml_consistency"
...
...
...
```

so far it works, ofc i cannot make it "work" without the new code hosted somewhere for the script to use
maybe i will try to figure out how to make jitpack with work kotlin scripts later..

krzema12:
Its enough for me, thanks :) I'll take care of merging it properly tomorrow. Thank you for the contribution! 🎉

krzema12 (approved):

krzema12:
@NikkyAI there are some conflicts - could you resolve them please?

NikkyAI:
> @NikkyAI there are some conflicts - could you resolve them please?

i see no conflicts ?

krzema12:
Merged manually.

PR #134, opened 2022-03-16
Add all remaining triggers

jmfayard:
Later need to add type-safe arguments for the remaining triggers, especially an enum for "types", but that will be backward compatible.

Change the way triggers are transformed to yaml, so that it can be used by the script-generator to transform to Kotlin

Add @Serializable so that it can be used later by the script-generator

### DIFF
--- a/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/BranchProtectionRule.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/BranchProtectionRule.kt
@@ -1,0 +1,9 @@
+package it.krzeminski.githubactions.domain.triggers
+
+import it.krzeminski.githubactions.dsl.CustomValue
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class BranchProtectionRule(
+    override val _customArguments: Map<String, CustomValue> = mapOf(),
+) : Trigger()
--- a/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/BranchProtectionRule.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/BranchProtectionRule.kt
@@ -1,9 +1,7 @@
 package it.krzeminski.githubactions.domain.triggers
 
 import it.krzeminski.githubactions.dsl.CustomValue
-import kotlinx.serialization.Serializable
 
-@Serializable
 data class BranchProtectionRule(
     override val _customArguments: Map<String, CustomValue> = mapOf(),
 ) : Trigger()
--- a/library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
@@ -56,10 +56,10 @@ private fun Job.toYaml() = buildString {
         }
     }
 
+    append(customArgumentsToYaml())
     appendLine("  steps:")
     append(steps.stepsToYaml().prependIndent("    "))
-    append(customArgumentsToYaml())
-}
+}.removeSuffix("\n")
 
 fun requireMatchesRegex(field: String, value: String, regex: Regex, url: String?) {
     require(regex.matchEntire(value) != null) {
--- a/library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
@@ -20,12 +20,6 @@ fun List<Job>.jobsToYaml(): String =
 
 private fun Job.toYaml() = buildString {
     val job = this@toYaml
-    requireMatchesRegex(
-        field = "Job.name",
-        value = job.name,
-        regex = Regex("[a-zA-Z_][a-zA-Z0-9_-]*"),
-        url = "https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#setting-an-id-for-a-job"
-    )
     appendLine("\"${job.name}\":")
     appendLine("  runs-on: \"${runsOn.toYaml()}\"")
 
@@ -56,16 +50,13 @@ private fun Job.toYaml() = buildString {
         }
     }
 
-    append(customArgumentsToYaml())
+    job.timeoutMinutes?.let { value ->
+        appendLine("  timeout-minutes: $value")
+    }
+
     appendLine("  steps:")
     append(steps.stepsToYaml().prependIndent("    "))
-}.removeSuffix("\n")
-
-fun requireMatchesRegex(field: String, value: String, regex: Regex, url: String?) {
-    require(regex.matchEntire(value) != null) {
-        val seeUrl = if (url.isNullOrBlank()) "" else "\nSee: $url"
-        """Invalid field ${field.replace('.', '(')}="$value") does not match regex: $regex$seeUrl"""
-    }
+    append(customArgumentsToYaml())
 }
 
 fun RunnerType.toYaml() =
--- a/library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
@@ -56,6 +56,7 @@ private fun Job.toYaml() = buildString {
 
     appendLine("  steps:")
     append(steps.stepsToYaml().prependIndent("    "))
+    if (_customArguments.isNotEmpty()) appendLine()
     append(customArgumentsToYaml())
 }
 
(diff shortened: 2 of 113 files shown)


krzema12:
There are several things mixed in this PR, let's start with having a separate PR for adding the triggers, without adding serialization stuff or adjusting custom YAML serialization logic.

jmfayard:
No problem for removing the kotlinx.serialization stuff, 
but I would rather not implement YAML serialization for 30 new wrappers just to revert it later because there is a better way that works both for serialization do YAML *and* Kotlin :)

jmfayard:
kotlinx.serialization removed from library module

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
```diff
@@ -2,77 +2,223 @@
...
 
-private fun PullRequestTarget.toYaml() = buildString {
-    appendLine("pull_request_target:")
-    printIfHasElements(this@toYaml.types.map(PullRequestTarget.Type::toYaml), "types")
-    printIfHasElements(this@toYaml.branches, "branches")
-    printIfHasElements(this@toYaml.branchesIgnore, "branches-ignore")
-    printIfHasElements(this@toYaml.paths, "paths")
-    printIfHasElements(this@toYaml.pathsIgnore, "paths-ignore")
-}.removeSuffix("\n")
+val triggerClassMap: List<Pair<String, KClass<*>>> = listOf(
```
krzema12, line 98: Observation: this property becomes a part of public API, this library makes use of it by searching by values, and there's some reflection going on here (in case of missing entry in this map, the code will fail in runtime). To be honest, seems shady and a bit hacky :raised_eyebrow: :smile: Let's try to find a better solution.

My goal is to keep this library high-quality and maintainable, while side projects like YAML-to-Kotlin converter should have as little impact on the library as possible. By exposing such semi-public API to be used by the converter, the library extends its hand to the converter sacrificing its own quality. I'd rather go the other way: first implement all the triggers as if the converter wasn't there, and then find the best approach so that the converter can do its job. It's also more expected for the converter to use reflection because that code will be less often used, and acts like an addition to the library.

What comes to my mind as something satisfactory for the converter, and preserving the right level of quality here in the library: having YAML name defined in each trigger class. Then the converter could use reflection to get all subclasses of `Trigger` and find such class that has the expected YAML name. It would also mean no new piece of public API for this library, just some internal fields that the library uses internally, and the converter happens to know about them and access them by reflection.

WDYT? I hope you get my point, and we can work something out :muscle:

library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
```diff
@@ -56,6 +56,7 @@ private fun Job.toYaml() = buildString {
 
     appendLine("  steps:")
     append(steps.stepsToYaml().prependIndent("    "))
+    if (_customArguments.isNotEmpty()) appendLine()
```
krzema12, line 59: Is this change somehow related to adding the triggers?

library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/BranchProtectionRule.kt
```diff
@@ -0,0 +1,7 @@
+package it.krzeminski.githubactions.domain.triggers
+
+import it.krzeminski.githubactions.dsl.CustomValue
+
+data class BranchProtectionRule(
```
krzema12, line 5: There at least two things mixed in this PR: adding all the triggers, and changing how they're serialized to YAML. The first part looks like something repetitive, a no-brainer that could go to a separate PR. This way we could simplify this PR. It's much easier for me to review several simpler PRs than a big one where several things are mixed - could you split it please? :pray:

Also, see my other PR where I suggest putting a YAML name for each trigger somewhere around class declaration. This way we may be able to satisfy both aspects at once.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
jmfayard: I fully understand the concern.

Can we declare a property inside `Trigger` to be implemented by each class? That would be enough for me.

Something like:

```kotlin
sealed class Trigger : HasCustomArguments {
    abstract val _triggerName: String
}

data class BranchProtectionRule(
    override val _customArguments: Map<String, CustomValue> = mapOf(),
) : Trigger() {
    override val _triggerName = "branch_protection_rule"
}
```

library/src/main/kotlin/it/krzeminski/githubactions/yaml/JobsToYaml.kt
jmfayard: Yes, I had to remove an `append("\n")` in `ToYaml.kt` and reintroduce it here

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
jmfayard: I fully agree with the concern.

In fact I had a look and I don't need it, I have suppresed it in a new commit

library/src/main/kotlin/it/krzeminski/githubactions/domain/triggers/BranchProtectionRule.kt
jmfayard: The part that changes how they are serialized is simple:
It refactors this imperative call to `buildString()` here
https://github.com/krzema12/github-actions-kotlin-dsl/blob/9cd86cf9864ecad897c5c00d47fc79468564e92d/library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt#L27-L35

to a more functional style
https://github.com/krzema12/github-actions-kotlin-dsl/blob/5f1483c12de92b0609cd4032302e45eaaa59517d/library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt#L181-L188

called by this

https://github.com/krzema12/github-actions-kotlin-dsl/blob/5f1483c12de92b0609cd4032302e45eaaa59517d/library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt#L135-L141

The parts that don't fit this pattern are handled here:

https://github.com/krzema12/github-actions-kotlin-dsl/blob/5f1483c12de92b0609cd4032302e45eaaa59517d/library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt#L135-L141

The hardest part was where to add `\n` or remove it, see your comment on this.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
```diff
@@ -2,90 +2,223 @@
...
-    printIfHasElements(this@toYaml.pathsIgnore, "paths-ignore")
-}.removeSuffix("\n")
+@Suppress("ComplexMethod")
+fun Trigger.toMap(): MapOfYaml {
+    return when (this) {
+        is Push -> toMap()
+        is PullRequest -> toMap()
+        is PullRequestTarget -> toMap()
+        is WorkflowDispatch -> emptyMap
+        is Schedule -> emptyMap
```
krzema12, line 64: Maybe `else -> emptyMap`?

krzema12 (changes requested):
Looks good in general, some nitpicking.

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
jmfayard: Done

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
```diff
@@ -2,90 +2,223 @@
...
+    is Watch -> "watch"
+    is WorkflowCall -> "workflow_call"
+    is WorkflowRun -> "workflow_run"
+}
+
+@Suppress("ComplexMethod")
+private fun Trigger.toAdditionalYaml(): String = when (this) {
+    is Schedule -> toAdditionalYaml()
+    is WorkflowDispatch -> toAdditionalYaml()
+    is Push -> ""
```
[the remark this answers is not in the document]
jmfayard: Done

library/src/main/kotlin/it/krzeminski/githubactions/yaml/TriggersToYaml.kt
```diff
@@ -2,90 +2,223 @@
...
 
-private fun PullRequest.Type.toYaml(): String = this.toSnakeCase()
+private val emptyMap = LinkedHashMap<String, List<String>?>()
 
-private fun Schedule.toYaml() = buildString {
-    appendLine("schedule:")
-    this@toYaml.triggers.forEach {
-        appendLine(" - cron: '${it.expression}'")
+@Suppress("ComplexMethod")
+fun Trigger.triggerName() = when (this) {
```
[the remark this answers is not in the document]
jmfayard: @krzema12 I did using Kotlin's `@RequireOptIn` feature which is made for that
[feat: annotate as @InternalGithubActionsApi methods meant for the scr…](https://github.com/krzema12/github-actions-kotlin-dsl/pull/134/commits/dbca26975b47ba30911996fb915a9c4ddc94f66d)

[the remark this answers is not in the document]
krzema12: Great idea! :)

krzema12 (approved):

PR #324, opened 2022-06-28
type-safe environment variables and secrets

jmfayard:
How to use: 

see tests

Links:

- Environment variables: https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables
- Default Environment variables: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

### DIFF
--- a/library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/Env.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/Env.kt
@@ -1,0 +1,169 @@
+package it.krzeminski.githubactions.dsl.expressions
+
+/**
+ * Environment variables: https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables
+ * Default Environment variables: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+ */
+object Env : ExpressionContext(_path = "env", map = MapFromLambda { key: String -> "\$$key" }) {
+    /** Always set to true. **/
+    val CI by map
+
+    /** The name of the action currently running, or the id of a step.
+     * For example, for an action, __repo-owner_name-of-action-repo.GitHub removes special characters,
+     * and uses the name __run when the current step runs a script without an id.
+     * If you use the same script or action more than once in the same job,
+     * the name will include a suffix that consists of the sequence number preceded by an underscore.
+     * For example, the first script you run will have the name __run, and the second script will be named __run_2.
+     * Similarly, the second invocation of actions/checkout will be actionscheckout2. **/
+    val GITHUB_ACTION by map
+
+    /** The path where an action is located. This property is only supported in composite actions.
+     * You can use this path to access files located in the same repository as the action.
+     * For example, /home/runner/work/_actions/repo-owner/name-of-action-repo/v1. **/
+    val GITHUB_ACTION_PATH by map
+
+    /** For a step executing an action, this is the owner and repository name of the action.
+     * For example, actions/checkout. **/
+    val GITHUB_ACTION_REPOSITORY by map
+
+    /** Always set to true when GitHub Actions is running the workflow.
+     * You can use this variable to differentiate when tests are being run locally or by GitHub Actions. **/
+    val GITHUB_ACTIONS by map
+
+    /** The name of the person or app that initiated the workflow.
+     * For example, octocat. **/
+    val GITHUB_ACTOR by map
+
+    /** Returns the API URL.
+     * For example: https://api.github.com. **/
+    val GITHUB_API_URL by map
+
+    /** The name of the base ref or target branch of the pull request in a workflow run.
+     * This is only set when the event that triggers a workflow run is either pull_request or pull_request_target.
+     * For example, main. **/
+    val GITHUB_BASE_REF by map
+
+    /** The path on the runner to the file that sets environment variables from workflow commands.
+     * This file is unique to the current step and changes for each step in a job.
+     * For example, /home/runner/work/_temp/_runner_file_commands/set_env_87406d6e-4979-4d42-98e1-3dab1f48b13a.
+     * For more information, see "Workflow commands for GitHub Actions." **/
+    val GITHUB_ENV by map
+
+    /** The name of the event that triggered the workflow. For example, workflow_dispatch. **/
+    val GITHUB_EVENT_NAME by map
+
+    /** The path to the file on the runner that contains the full event webhook payload.
+     * For example, /github/workflow/event.json. **/
+    val GITHUB_EVENT_PATH by map
+
+    /** Returns the GraphQL API URL. For example: https://api.github.com/graphql. **/
+    val GITHUB_GRAPHQL_URL by map
+
+    /** The head ref or source branch of the pull request in a workflow run.
+     * This property is only set when the event that triggers
+     * a workflow run is either pull_request or pull_request_target.
+     * For example, feature-branch-1. **/
+    val GITHUB_HEAD_REF by map
+
+    /** The job_id of the current job.
+     * For example, greeting_job. **/
+    val GITHUB_JOB by map
+
+    /** The path on the runner to the file that sets system PATH variables from workflow commands.
+     * This file is unique to the current step and changes for each step in a job.
+     * For example, /home/runner/work/_temp/_runner_file_commands/add_path_899b9445-ad4a-400c-aa89-249f18632cf5.
+     * For more information, see "Workflow commands for GitHub Actions." **/
+    val GITHUB_PATH by map
+
+    /** The branch or tag ref that triggered the workflow run.
+     * For branches this is the format refs/heads/<branch_name>,
+     * for tags it is refs/tags/<tag_name>,
+     * and for pull requests it is refs/pull/<pr_number>/merge.
+     * This variable is only set if a branch or tag is available for the event type.
+     * For example, refs/heads/feature-branch-1. **/
+    val GITHUB_REF by map
+
+    /** The branch or tag name that triggered the workflow run.
+     * For example, feature-branch-1. **/
+    val GITHUB_REF_NAME by map
+
+    /** true if branch protections are configured for the ref that triggered the workflow run. **/
+    val GITHUB_REF_PROTECTED by map
+
+    /** The type of ref that triggered the workflow run.
+     * Valid values are branch or tag. **/
+    val GITHUB_REF_TYPE by map
+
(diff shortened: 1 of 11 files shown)


library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
```diff
@@ -1,11 +1,11 @@
 package it.krzeminski.githubactions.dsl.expressions
 
 open class ExpressionContext(
-    private val path: String,
-    val propertyToExprPath: Map<String, String> = MapFromLambda { propertyName -> "$path.$propertyName" }
-)
+    internal val _path: String,
```
jmfayard, line 4: `path`  doesn't work, there are contexts with a property called `path`

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
```diff
@@ -1,11 +1,11 @@
...
+    val map: Map<String, String> = MapFromLambda { propertyName -> "$_path.$propertyName" }
```
jmfayard, line 5: we can revert to `propertyToExprPath` but that feels weird to me

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
```diff
@@ -1,11 +1,11 @@
...
+): Map<String, String> by map
 
-private class MapFromLambda<T>(val operation: (String) -> T) : Map<String, T> by emptyMap() {
+internal class MapFromLambda<T>(val operation: (String) -> T) : Map<String, T> by emptyMap() {
```
jmfayard, line 8: If it's private, I can't use it in `Env` and `Secrets`

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
```diff
@@ -1,11 +1,11 @@
 package it.krzeminski.githubactions.dsl.expressions
 
 open class ExpressionContext(
-    private val path: String,
-    val propertyToExprPath: Map<String, String> = MapFromLambda { propertyName -> "$path.$propertyName" }
-)
+    internal val _path: String,
+    val map: Map<String, String> = MapFromLambda { propertyName -> "$_path.$propertyName" }
+): Map<String, String> by map
```
jmfayard, line 6: `val TOKEN by Env` instead of `by Env.map`

`val SUPER_SECRET by Secrets` instead of `by Secrets.map`

library/src/test/resources/expected/workflow-expr-typesafe.yml
```diff
@@ -10,9 +10,25 @@ on:
...
+      - id: step-2
+        name: Custom environment variable
+        env:
+          FIRST_NAME: Patrick
+        run: echo ${{ $GREETING }} ${{ $FIRST_NAME }}
+      - id: step-3
+        name: Encrypted secret
+        env:
+          SECRET: ${{ secrets.SUPER_SECRET }}
+          TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jmfayard, line 30: Without the change to `fun LinkedHashMap<String, String>.toYaml()` that would be:

`$SECRET: ${{ secrets.SUPER_SECRET }}`

That the one place where we don't want this behavior

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/Env.kt
```diff
@@ -0,0 +1,169 @@
+package it.krzeminski.githubactions.dsl.expressions
```
krzema12, line 1: Should be in 
```suggestion
package it.krzeminski.githubactions.dsl.expressions.contexts
```

`Secrets` as well.

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
krzema12: True!

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
krzema12: I wanted to come up with anything more meaningful than `map`, but I agree that `propertyToExprPath` does look weird. Maybe simply `contextProperty`?

Please also revert the renaming in this commit and let's rename it in a separate one, to avoid noise.

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/Secrets.kt
```diff
@@ -0,0 +1,18 @@
+package it.krzeminski.githubactions.dsl.expressions
+
+/**
+ * Encrypted secrets
+ * Encrypted secrets allow you to store sensitive information in your organization,
+ * repository, or repository environments.
+ *
+ * https://docs.github.com/en/actions/security-guides/encrypted-secrets
+ */
+object Secrets : ExpressionContext("secrets", MapFromLambda { key: String -> "secrets.$key" }) {
```
krzema12, line 10: Why is it needed to pass the second constructor argument? Won't the default work here?

Also, `SecretsContext` for consistency.

library/src/test/kotlin/it/krzeminski/githubactions/IntegrationTest.kt
```diff
@@ -704,11 +709,39 @@ class IntegrationTest : FunSpec({
             on = listOf(Push()),
             sourceFile = Path.of("ExprIntegrationTest.kt"),
         ) {
+            val GREETING by Env
+            val FIRST_NAME by Env
+            val SECRET by Env
+            val TOKEN by Env
+            val SUPER_SECRET by Secrets
```
krzema12, line 716: Could we somehow differentiate how custom values are defined? Now it's not clear and one may think that `SUPER_SECRET` is a part of GitHub-provided context. How about:

```suggestion
            val SUPER_SECRET by Secrets.custom
```
? Any thoughts?

library/src/main/kotlin/it/krzeminski/githubactions/yaml/Utils.kt
```diff
@@ -1,7 +1,9 @@
 package it.krzeminski.githubactions.yaml
 
 fun LinkedHashMap<String, String>.toYaml() =
-    map { (key, value) ->
+    map { (rawKey, value) ->
+        // Needed for type-safe environment variables. See test("Executing workflow with type-safe expressions")
+        val key = rawKey.removePrefix("\$")
```
krzema12, line 6: Can it be encapsulated in `Env`, the logic inside `MapFromLambda`?

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/Env.kt
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Environment variables: https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables
+ * Default Environment variables: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+ */
+object Env : ExpressionContext(_path = "env", map = MapFromLambda { key: String -> "\$$key" }) {
```
krzema12, line 7:
```suggestion
object EnvContext : ExpressionContext(_path = "env", map = MapFromLambda { key: String -> "\$$key" }) {
```

library/src/test/resources/expected/workflow-expr-typesafe.yml
krzema12: Let's discuss another approach where to handle it - I added a comment in `toYaml()`.

krzema12 (changes requested):

krzema12:
Sorry for failing "Check if wrappers up to date" check - as we have more actions, it happens more and more frequently. We have fixing it on the radar as https://github.com/krzema12/github-actions-kotlin-dsl/issues/208.  For the purpose of this PR you can ignore it, I'll handle it.

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/Secrets.kt
```diff
@@ -0,0 +1,18 @@
+package it.krzeminski.githubactions.dsl.expressions
```
krzema12, line 1:
```suggestion
package it.krzeminski.githubactions.dsl.expressions.contexts
```

library/src/main/kotlin/it/krzeminski/githubactions/yaml/Utils.kt
jmfayard: The logic inside `EnvContext` is to add the dollar, because without it, instead you would have to write this which is error-prone

```kotlin
run(
    name = "Default environment variable",
-    command = "action=${Contexts.env.GITHUB_ACTION} repo=${Contexts.env.GITHUB_REPOSITORY}",
+   command = "action=\$${Contexts.env.GITHUB_ACTION} repo=\$${Contexts.env.GITHUB_REPOSITORY}",
)
```

The reason I have put it there is that modifying all places where we can have a `env` looks like a lot of work
github-actions-kotlin-dsl_–_Step_kt__github-actions-kotlin-dsl_library_main_

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/ExpressionContext.kt
jmfayard: done

library/src/main/kotlin/it/krzeminski/githubactions/yaml/Utils.kt
krzema12: Ok, I think I get it now. It depends on the context if we want to have the dollar in resulting YAML and here's the best place to express this knowledge.

library/src/test/kotlin/it/krzeminski/githubactions/IntegrationTest.kt
krzema12: Ok I wasn't aware of this naming convention :) Here I pointed it out looking purely on library's API. Let's leave it as is for now and see how it works in practice.

krzema12 (approved):

PR #325, opened 2022-06-29
type-safe functions

jmfayard:
This one should be pretty straight forward

Reference: https://docs.github.com/en/actions/learn-github-actions/expressions#functions

### DIFF
--- a/library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/contexts/FunctionsContext.kt
+++ b/library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/contexts/FunctionsContext.kt
@@ -1,5 +1,7 @@
 package it.krzeminski.githubactions.dsl.expressions.contexts
 
+import kotlin.text.Typography.quote
+
 /**
  * GitHub offers a set of built-in functions that you can use in expressions.
  * Some functions cast values to a string to perform comparisons.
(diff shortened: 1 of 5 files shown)


library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/contexts/FunctionsContext.kt
```diff
@@ -0,0 +1,48 @@
...
+ * GitHub offers a set of built-in functions that you can use in expressions.
+ * Some functions cast values to a string to perform comparisons.
+ * GitHub casts data types to a string using these conversions:
+ *
+ * https://docs.github.com/en/actions/learn-github-actions/expressions#functions
+ */
+@Suppress("FunctionOnlyReturningConstant", "TooManyFunctions")
+open class FunctionsContext {
+    private fun formatArgs(vararg args: String, quote: Boolean) =
+        args.joinToString(prefix = "(", postfix = ")") { if (quote) "'$it'" else it }
```
krzema12, line 13: We should probably escape `'` in the args, otherwise it will produce an invalid YAML.

krzema12 (changes requested):
Wow, love it! Just one thing to address.

library/src/main/kotlin/it/krzeminski/githubactions/dsl/expressions/contexts/FunctionsContext.kt
jmfayard: How exactly?
Escaping as `\'` seems like a syntax error in YAML.
Here "if3 and if4" are wrong while the others are correct.
Online_YAML_Parser

krzema12: What I meant is that we should escape `'` if the user passes it in `args`:

```
join("some 'quoted' string", "another string")
```

It should be something like

```
join('some \'quoted\' string', 'another string')
```

instead of

```
join('some 'quoted' string', 'another string')
```

jmfayard: People don't seem to escape `'`  in their workflows

https://github.com/search?q=startsWith+filename%3A.github%2Fworkflows%2F*.yml

krzema12: ktlint is sad :)

jmfayard: Sorry I had the fix but forgot to git push --force it :)

PR #682, opened 2023-02-13
feat(library): allow setting workflow env values from within workflow builder

Vampire:


Vampire:
Misinterpreted the need to set `env` on workflow dynamically.
If you think it would still make sense, please reopen.

PR #686, opened 2023-02-16
docs: Extract snippets from documentation to tested files and fix invalid examples (#195)

Vampire:
Fixes: #195

Contentwise changes in the snippets:
- removed some trailing whitespace
- added an inline code highlighting
- replaced deprecated action with its current version
- fixed all the ktlint findings in the snippets

Everything else is currently identical to the previous outcome.

### DIFF
--- a/library/build.gradle.kts
+++ b/library/build.gradle.kts
@@ -71,6 +71,10 @@ val validateDuplicatedVersion by tasks.creating<Task> {
             project.rootDir.resolve("script-generator/src/main/kotlin/it/krzeminski/githubactions/scriptgenerator/Version.kt").readText()
                 .contains("val LIBRARY_VERSION = \"$version\"")
         ) { "Library version stated in script-generator/.../Version.kt should be equal to $version!" }
+        require(
+            project.file("src/test/kotlin/it/krzeminski/githubactions/docsnippets/GettingStartedSnippets.kt").readText()
+                .contains("\"it.krzeminski:github-actions-kotlin-dsl:$version\"")
+        ) { "Library version stated in library/src/test/.../GettingStarted.kt should be equal to $version!" }
     }
 }
 
--- a/library/src/test/kotlin/it/krzeminski/githubactions/docsnippets/GettingStartedSnippets.kt
+++ b/library/src/test/kotlin/it/krzeminski/githubactions/docsnippets/GettingStartedSnippets.kt
@@ -1,0 +1,45 @@
+package it.krzeminski.githubactions.docsnippets
+
+/* ktlint-disable import-ordering */
+import io.kotest.core.spec.style.FunSpec
+// --8<-- [start:gettingStarted2]
+import it.krzeminski.githubactions.actions.actions.CheckoutV3
+import it.krzeminski.githubactions.domain.RunnerType.UbuntuLatest
+import it.krzeminski.githubactions.domain.triggers.Push
+import it.krzeminski.githubactions.dsl.workflow
+import it.krzeminski.githubactions.yaml.toYaml
+// --8<-- [end:gettingStarted2]
+import java.io.File
+/* ktlint-enable import-ordering */
+
+class GettingStartedSnippets : FunSpec({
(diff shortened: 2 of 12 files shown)


krzema12:
I generally prefer making side changes as separate PRs, but I can take it from here and split according to what my internal purist suggests ;)

Vampire:
Yeah, I'm usually the same, but the deprecated class would not have compiled without a suppression, or I could have made the change first in the inline version.
I can surely split it further if you prefer.

krzema12:
> or I could have made the change first in the inline version.

Yes, it would be cool to update the action version first (also remove any trailing whitespace you noticed), e. g. as separate commits in a single PR.

About syntax highlighting, I thought it's already enabled. Could you show an example?

Vampire:
I've split it to separate commits in this PR, if that's ok for you :-)

krzema12:
Sure!

library/src/test/kotlin/it/krzeminski/githubactions/docsnippets/GettingStartedSnippets.kt
```diff
@@ -0,0 +1,41 @@
...
+// --8<-- [end:gettingStarted2]
+import java.io.File
+
+class GettingStartedSnippets : FunSpec({
+    test("gettingStarted") {
+/*
+// --8<-- [start:gettingStarted1]
+#!/usr/bin/env kotlin
+
+@file:DependsOn("it.krzeminski:github-actions-kotlin-dsl:0.36.0")
```
krzema12, line 19: I see that the lib version doesn't come from mkdocs placeholder. What happens if we bump the version? Is there a risk we'll forget to update it here? Maybe the placeholder works here?

library/build.gradle.kts
```diff
@@ -61,6 +61,17 @@ tasks.formatKotlinMain {
     kotlinterConfig()
 }
 
+fun ConfigurableKtLintTask.kotlinterTestConfig() {
+    exclude { it.file.invariantSeparatorsPath.contains("/docsnippets/") }
```
krzema12, line 65: Instead of excluding all snippet files, have you tried disabling just the checks that will usually fail, like for the indent? I think it's still beneficial to run the linter on the examples for the docs.

library/src/test/kotlin/it/krzeminski/githubactions/docsnippets/GettingStartedSnippets.kt
krzema12: Ok, now I see the check in Gradle! So the mkdocs placeholder doesn't work here?

Vampire: I tried with the placeholder first, as I copied stuff 1:1 and just surrounded it to be compilable and runnable.
It was not replaced unfortunately.
Maybe somthing to report to that extension, I don't know.
Risk to forget updating is 0, as I added it to the `validateDuplicatedVersion` task where you also verify the explicit versions in `mkdocs.yml` and `Version.kt` for consistency so that would also fail for forgetting an update there.

An alternative would be to have that line in the docs directly instead of the snippet, but I thought it would be preferable to keep the snippet code together.

library/build.gradle.kts
Vampire: No, I did not, I just disabled it right away as I knew how to do it :-)

Vampire: And because I didn't want to fix all the violations in the examples but keep them unchanged.
Especially those super-ugly trailing commas. :-D

krzema12: Got it, we can take it out of scope of this PR :) I might tackle it one day

Vampire: Ok, I now added another prepare commit that first fixes all the ktlint findings in the snippets and targetedly only suppress the indentation rule where appropriate and at one place the import ordering rule.

krzema12 (approved):

krzema12:
image

PR #1598, opened 2024-08-19
feat(server): allow to refresh generated artifacts on demand

Vampire:


Vampire:
* **#1600** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1599** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1598** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1598?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
```diff
@@ -54,6 +55,12 @@ fun main() {
                 artifact(bindingsCache)
             }
 
+            route("/refresh") {
```
krzema12, line 58: I initially thought that this route just refreshes the cache for a given action, but it actually acts as an alternative repository URL for the scripts. I think it's pretty unsafe. Why: assuming people's good will, someone may just forget to remove the `/refresh` URL suffix, and then each workflow run will result in repopulating cache for all used actions which will put extra load on the server.

What I'd propose instead is to make this endpoint just refresh artifacts for a given action, and return a simple status like `OK` to accompany a HTTP 200 status code. Then it's harder to misuse it, and still possible to refresh desired actions.

Vampire:
Actually, the main target-use-case was developing changes in the library using hot-swapped changes and then just getting the new result, also putting it to the cache for normal requests.

That it could also be handy for nuking the production cache e.g. if you know there was a new release or changed typings or similar is mainly a positive side-effect.

I changed it now so that for a normal user / production deployment only OK is answered, but you can change it to respond using an environment variable, this can then e.g. be settable when running from IDE.

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
Vampire: Actually, the main target-use-case was developing changes in the library using hot-swapped changes and then just getting the new result, also putting it to the cache for normal requests.

That it could also be handy for nuking the production cache e.g. if you know there was a new release or changed typings or similar is mainly a positive side-effect.

I changed it now so that for a normal user / production deployment only OK is answered, but you can change it to respond using an environment variable, this can then e.g. be settable when running from IDE.

.github/workflows/test-script-consuming-jit-bindings-refresh.main.do-not-compile.kts
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env kotlin
+@file:Repository("https://repo.maven.apache.org/maven2/")
+@file:DependsOn("io.github.typesafegithub:github-workflows-kt:1.13.0")
+
+@file:Repository("http://localhost:8080/refresh/")
```
krzema12, line 5: I think this script doesn't make much sense because we now disallow using the new endpoint as a repo (at least by default), and this script isn't used in `.github/workflows/bindings-server.main.kts` anyway. Unless you want to test if `GWKT_DELIVER_ON_REFRESH`, but I won't insist on it.

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
```diff
@@ -58,15 +59,30 @@ fun main() {
...
+                route("{owner}/{name}/{version}/{file}") {
+                    artifact(bindingsCache, refresh = true)
+                }
+
+                route("{owner}/{name}/{file}") {
+                    metadata(refresh = true)
+                }
```
krzema12, line 69: We currently don't cache whatever is provided by this endpoint, only `{owner}/{name}/{version}/{file}`, so there's no point in refreshing it. However, I see that we want to have in case `GWKT_DELIVER_ON_REFRESH` is enabled. If so, we should guard returning a value from this route as well. See line 98, it's not guarded with `if (refresh && !deliverOnRefreshRoute)`.

krzema12 (changes requested):

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
Vampire: How about the very first `if` in `metadata` on line 81?!?

.github/workflows/test-script-consuming-jit-bindings-refresh.main.do-not-compile.kts
Vampire: Yeah, definitely, that's just a left-over from the unguarded state.
Now it is removed.

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
krzema12: Right, so we won't return the value - my bad, I misread because my brain was looking for returning "OK" in this case, like we do for the other endpoint. Returning a 404 is a bit confusing. Could we return 200 and OK?

krzema12: Sorry, OK doesn't make sense here. Not found is fine. I'm tired today

krzema12 (approved):

PR #1602, opened 2024-08-19
feat(server): make it possible to POST custom typings for testing during typing creation

Vampire:
With this you can for example do `curl --data-binary @action-types.yml https://bindings.krzeminski.it/pbrisbin/setup-tool-action/v2/setup-tool-action-v2.pom`
then get as result coordinates containing a UUID in the group field which you can then use for an hour (or from the Maven Local cache thereafter) to test with
the supplied typing.


Vampire:
* **#1602** <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/1602?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2298** <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2298?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#2308](https://github.com/typesafegithub/github-workflows-kt/pull/2308) <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2308?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2309](https://github.com/typesafegithub/github-workflows-kt/pull/2309) <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2309?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
```diff
@@ -127,39 +130,58 @@ private fun Route.artifact(
...
         }
         if (file in bindingArtifacts) {
-            call.respondText("Exists", status = HttpStatusCode.OK)
+            call.respondText(text = "Exists", status = HttpStatusCode.OK)
         } else {
             call.respondText(text = "Not found", status = HttpStatusCode.NotFound)
         }
     }
+
+    post {
```
krzema12, line 143: I'm afraid such feature on the shared server could easily let someone destabilize the server by using the fact that we have anchors & aliases enabled, and then use the [billion laughs attack](https://en.wikipedia.org/wiki/Billion_laughs_attack#Variations). I know it's already possible by creating an action with malicious action-types.yml and requesting it, and I think we eventually will have to do something about it (e.g. introduce some timeout for parsing the YAML, if it would be sufficient), but I'd rather not provide another entry point that makes it even easier to induce this problem.

That being said, I see some value in being able to test this way. Do we have any other way? I can think of starting a private instance of the server locally, and if you have public IP, things should work, but it would be a bit cumbersome. Or maybe we should create a CLI that would let someone generate the binding with provided typings locally, and just test with Maven Local?

krzema12: See [maven-binding-builder](https://github.com/typesafegithub/github-workflows-kt/tree/main/maven-binding-builder), it's a module that is currently used only in the bindings server, but I extracted it to contain the logic of building the bindings, and it's perfectly reusable for a potential CLI to generate such bindings locally and put in Maven local (yesterday I wanted to merge this module with the server, but it turns out that this split may be useful after all 😄).

Vampire: Hm, if there are such attacks, I think it is better to add some protection if necessary to all paths that might hit it instead of "not adding another entrypoint".

Of course there are other ways, you do not even need a public IP unless you want to test on GHA itself. But for the intended use-case you can just start the server locally and execute the script with that repository. But the point is, that someone like you or me would do that. But an arbitrary action author would not. To spread the word, it has to be convenient and that was my main goal. A CLI could also be created, but is imho not as convenient as just having it like proposed here.

But let's shift this a bit, and I'll maybe play a bit with attacking the laughs if I find some minutes to see what happens or could be done.

Vampire: SnakeYAML KMP actually is protected against billion laughs attack, both `Compose` and `Load` have protective code. KAML on the other hand does not have protection as it uses below-low-level API and handles resolution itself.

But luckily enough, there is a pull-request that adds billion-laughs protection to KAML: https://github.com/charleskorn/kaml/pull/620 :-)

Vampire: And in the meantime #1662 :-)

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
```diff
@@ -160,41 +164,71 @@ private fun Route.artifact(
...
+        val owner = "${call.parameters["owner"]}__types__${randomUUID()}"
+        val name = call.parameters["name"]!!
+        val version = call.parameters["version"]!!
+        val types = call.receiveText()
+        runCatching {
+            Load().loadOne(types)
```
krzema12, line 183: How about making `toBindingArtifacts` communicate such case via its returned value? Then we'd avoid adding a direct dependency on snakeyam-engine-kmp, and generally having parsing these encapsulated in the function (not mentioning it's an abstraction leak because it's not realistic to have anything else than YAML here).

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypesProviding.kt
```diff
@@ -23,10 +24,12 @@ private val logger = logger { }
 internal fun ActionCoords.provideTypes(
     metadataRevision: MetadataRevision,
     fetchUri: (URI) -> String = ::fetchUri,
+    types: String? = null,
```
krzema12, line 27: I think we can make it better - adding `types` looks like a carveout in this function, and implies some hacks across the project like adding `typesUuid` to `ActionCoords`. I'd expect that just parsing of the YAML is triggered if the types as string are provided. I'll think about how to structure it better.

docs/user-guide/using-actions.md
```diff
@@ -23,7 +23,7 @@ For every action, a binding will be generated. However, some less popular action
 their inputs, so by default all inputs are of type `String`, have the suffix `_Untyped`, and additionally the class
 name will have an `_Untyped` suffix. The nullability of the inputs will be according to their required status.
 
-There are two ways of configuring typings:
```
krzema12, line 26: Looks funny :P but I'd separate configuring typings from testing them.

krzema12 (commented):
Did the next pass, not fully reviewed. Moderately actionable on you.

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
Vampire: Actually, I don't really agree.
The value is coming from the HTTP request, so is user input.
User input should always be considered to be not what you expect and be validated.

The body is an "argument" to the "API" and the "API" defines this has to be valid YAML.
So imho it is the right thing to explicitly validate the body to be valid YAML and also do that as soon as possible, not implicitly after several method calls two modules deep.

Also imho doing it that way would be the abstraction leak.
Because currently the place that parses the YAML is not aware that it is run behind an HTTP API.
Doing this change would need to make it aware that it is run behind an HTTP request, because it has to differentiate whether the types are coming as "API argument" so that it can result in "this is an unprocessable entity user error". If any of the other YAMLs has a parsing error, this is not a 4xx user error, as the user calling the HTTP API did not make an error, but it is a 5xx error as something on the server side is not working properly.

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypesProviding.kt
Vampire: What do you mean by "carveout"?

It says "give me the types, here is the YAML to use, if I don't give one explicitly default to requesting them from the action or catalog". For me this is not a "carveout" unless I don't get what you mean.

I also do not really see what "hacks" you mean. The "typesUuid" for me is not a hack but a generated part of the coordinates. It also is part of the coordinates, it is just split once and then stored into a separate field.

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/Main.kt
```diff
@@ -51,4 +51,4 @@ fun main() {
 
 val deliverOnRefreshRoute = System.getenv("GWKT_DELIVER_ON_REFRESH").toBoolean()
 
-suspend fun ApplicationCall.respondNotFound() = respondText("Not found", status = HttpStatusCode.NotFound)
+suspend fun ApplicationCall.respondNotFound() = respondText(text = "Not found", status = HttpStatusCode.NotFound)
```
krzema12, line 54: Such changes are fine, but they artificially make this PR bigger, scaring me off from reviewing it 😜 Let's either extract such unrelated ones to a separate PR(s), or just remove from this PR.

Vampire: Yeah, by 2 :-D
Sorry, I know you and thought I pulled them out.
Before the upstream refactoring they were all in the same file.

PR #1612, opened 2024-08-19
fix(abg): use the release date of the latest version for lastUpdated

Vampire:
This makes the metadata file content more stable and allows to provide a checksum,
as Kotlin for example is quite unhappy if it finds no checksum or cannot successfully verify it.

Without this every request would produce a different file and thus also a different checksum file,
so the checksum would never match the file contents.

PR #1611, opened 2024-08-19
feat: allow using version ranges to cope with stale Maven Local cache

Vampire:
With these changes you can for example depend on `actions:cache__save___major:[v4,v5)"` and get the latest v4 action with latest v4 typing by resolving to a concrete v4.x.y version but the output still contains just `v4`.
The same for `actions:cache__save___minor:[v4.0,v4.1)`, to get the latest v4.0 but in the output just have `v4.0`.

PR #1608, opened 2024-08-19
fix(server): use correct coordinates and file names in generated GMM file

Vampire:

PR #1603, opened 2024-08-19
feat(abg)!: provide untyped fields always and typed ones if type available

Vampire:
Fixes #1585

I have to confess, that I did not check every change in the unit test workflow files,
so they should probably be looked again closely. But I think it should have done the right thing.
Or at least how I envisioned it.

no typing =>
- only `_Untyped` class with `_Untyped` properties
- nullability according to required status

full typing =>
- same `_Untyped` class as above
- but marked as `@Deprecated` with a replace hint for the typed class (works in the IDE),
- additionally a typed class with each property as typed and untyped version
- all properties nullable, with runtime checks that
  - not the typed and untyped is set both for one property
  - one of the two is set for required properties

partly typing =>
- same `_Untyped` and typed class as above
- for inputs that have no type declared only the `_Untyped` property is added
- the properties that are only untyped have the nullability according to required status, also in the typed class

PR #1615, opened 2024-08-20
perf(abg): create JARs only when they are actually requested

Vampire:
In some cases, only e.g. a POM is necessary to fetch for a given action. Thanks to this change, no Kotlin compilation or JAR creation will happen when requesting an artifact that doesn't require it.

### DIFF
--- a/maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/JarBuilding.kt
+++ b/maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/JarBuilding.kt
@@ -22,8 +22,8 @@ import kotlin.io.path.div
 import kotlin.io.path.writeText
 
 internal data class Jars(
-    val mainJar: ByteArray,
-    val sourcesJar: ByteArray,
+    val mainJar: () -> ByteArray,
+    val sourcesJar: () -> ByteArray,
 )
 
 internal fun buildJars(
@@ -35,20 +35,28 @@ internal fun buildJars(
         generateBinding(owner = owner, name = name, version = version).also {
             if (it.isEmpty()) return null
         }
-    val (sourceFilePaths, compilationInputDir) = binding.prepareDirectoryWithSources()
 
-    val pathWithJarContents = compileBinding(sourceFilePaths = sourceFilePaths)
-    val mainJarByteArrayOutputStream = ByteArrayOutputStream()
-    mainJarByteArrayOutputStream.createZipFile(pathWithJarContents)
-    pathWithJarContents.toFile().deleteRecursively()
+    val mainJar by lazy {
+        val (sourceFilePaths, compilationInputDir) = binding.prepareDirectoryWithSources()
+        val pathWithJarContents = compileBinding(sourceFilePaths = sourceFilePaths)
+        val mainJarByteArrayOutputStream = ByteArrayOutputStream()
+        mainJarByteArrayOutputStream.createZipFile(pathWithJarContents)
+        pathWithJarContents.toFile().deleteRecursively()
+        compilationInputDir.toFile().deleteRecursively()
+        mainJarByteArrayOutputStream.toByteArray()
+    }
 
-    val sourcesJarByteArrayOutputStream = ByteArrayOutputStream()
-    sourcesJarByteArrayOutputStream.createZipFile(compilationInputDir)
-    compilationInputDir.toFile().deleteRecursively()
+    val sourcesJar by lazy {
+        val (_, compilationInputDir) = binding.prepareDirectoryWithSources()
+        val sourcesJarByteArrayOutputStream = ByteArrayOutputStream()
+        sourcesJarByteArrayOutputStream.createZipFile(compilationInputDir)
+        compilationInputDir.toFile().deleteRecursively()
+        sourcesJarByteArrayOutputStream.toByteArray()
+    }
 
     return Jars(
-        mainJar = mainJarByteArrayOutputStream.toByteArray(),
-        sourcesJar = sourcesJarByteArrayOutputStream.toByteArray(),
+        mainJar = { mainJar },
+        sourcesJar = { sourcesJar },
     )
 }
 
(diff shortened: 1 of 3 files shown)


Vampire:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1615?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1616** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1615** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1614** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1613** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1613?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1612** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1611** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1610** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1609** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1609?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1608** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1608?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1603** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1607** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1602** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1601** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1601?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1600** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1599** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1598** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1598?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/JarBuilding.kt
```diff
@@ -22,27 +22,34 @@ import kotlin.io.path.div
...
     val binding =
         generateBinding(metadataRevision = NewestForVersion, types = types).also {
             if (it.isEmpty()) return null
         }
-    val (sourceFilePaths, compilationInputDir) = binding.prepareDirectoryWithSources()
 
-    val pathWithJarContents = compileBinding(sourceFilePaths = sourceFilePaths)
-    val mainJarByteArrayOutputStream = ByteArrayOutputStream()
-    mainJarByteArrayOutputStream.createZipFile(pathWithJarContents)
+    val mainJar by lazy {
```
krzema12, line 35: The goal of creating all artifacts for a given action version at once was to make them consistent. So for example, at moment #1 I request just the compiled JAR, and at moment #2 I request the sources JAR. I want to be sure that they are consistent with each other, and not that the action owner introduces a change in the meantime. That's why I'd like to stay with how things work now. Unless I don't get the point and you explain it 🙂

Vampire: You don't get the point and I'll explain it. :-)

With my changes the same is still guaranteed.

The information necessary from the typing catalog and the action is retrieved on `generateBinding` which is where the information is collected and transformed to `ActionBinding` instances.
This call is made with the first request, it is needed for any artifact, as it is also needed to determine 404 response for example.

Even the source code is already constructed as string in the `ActionBinding` instances.

All this is quite fast.

When requesting one of the jars, then just this existing result that is part of the cache as the `by lazy` implementation uses the `binding` variable is written to files on disk, for the code jar compiled, packaged into a jar, and then delivered, preserving this result for later requests due to the `by lazy`.

This way the time-consuming tasks of disk IO, compilation, and jar compression are delayed and optional.


In the current binding server and when used from a workflow script, this would just be a minor performance optimization, as usually when the metadata was retrieved, the jar will also be retrieved. On the other hand, creating the source jar is usually not necessary.

When a dependency update bot retrieves the metadata, and it usually will do that to find the url to determine release notes for example, then it does not have to wait for the jar to be created needlessly, but quickly gets the metadata. Also here, typically the bot will create a PR and typically that PR will cause an action to be triggered, that typically then will need the code jar, but also not the source jar.

But why I actually looked into this was, because of the version-range support. It works fine without this, but for whatever reason the dependency resolution engine used by the main.kts scripts does not simply get the metadata file and determines the version from that, but it gets the poms of all the possible versions. And without this PR this means that even though the resolution engine just needs the poms, all the source and binary jars are created and never requested, also making the script execution quite slow as it has to wait for all the jars to be created needlessly.

krzema12: Thanks for this explanation! Probably worth to summarize it in 1-2 words in the PR description, I'll do it.
I also did some simple testing - before this change, requesting a POM from a server running locally takes 1.3-1.4 s, and after this change: 0.9-1.0 s. I think it makes sense to sacrifice a bit of readability to get this performance gain. Thanks!

krzema12 (approved):

PR #1646, opened 2024-09-08
refactor: unify content of ActionCoords name

Vampire:
Before these changes, `ActionCoords#name` sometimes had the actual action coordinates, so for example `some-name__some-sub`, sometimes it had the replaced `some-name/some-sub`.
This is not good, as you never know what you get and also led to certain bugs where wrong things are written to metadata files for example.
This changes unifies the situation, so that the `name` only contains the repository name and the `path` is in a separate property with slashes.
All other manipulation is then done where needed.

### DIFF
--- a/action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/domain/ActionCoords.kt
+++ b/action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/domain/ActionCoords.kt
@@ -4,37 +4,39 @@ public data class ActionCoords(
     val owner: String,
     val name: String,
     val version: String,
+    val path: String? = null,
 )
 
 /**
  * A top-level action is an action with its `action.y(a)ml` file in the repository root, as opposed to actions stored
  * in subdirectories.
  */
-public val ActionCoords.isTopLevel: Boolean get() = "/" !in name
+public val ActionCoords.isTopLevel: Boolean get() = path == null
 
-public val ActionCoords.prettyPrint: String get() = "$owner/$name@$version"
-
-/**
- * For most actions, it's the same as [ActionCoords.name].
- * For actions that aren't executed from the root of the repo, it returns the repo name.
- */
-public val ActionCoords.repoName: String get() =
-    name.substringBefore("/")
+public val ActionCoords.prettyPrint: String get() = "$owner/$fullName@$version"
 
 /**
  * For most actions, it's empty.
  * For actions that aren't executed from the root of the repo, it returns the path relative to the repo root where the
+ * action lives, starting with a slash.
+ */
+public val ActionCoords.subName: String get() = path?.let { "/$path" } ?: ""
+
+/**
+ * For most actions, it's equal to [ActionCoords.name].
+ * For actions that aren't executed from the root of the repo, it returns the path starting with the repo root where the
  * action lives.
  */
-public val ActionCoords.subName: String get() =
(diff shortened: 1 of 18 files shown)


Vampire:
* **#1646** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1646?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1602](https://github.com/typesafegithub/github-workflows-kt/pull/1602) <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1609](https://github.com/typesafegithub/github-workflows-kt/pull/1609) <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1609?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈
* `main`

<h2></h2>

krzema12 (commented):
There's a lot going on here - I presume that you have a certain principal idea in your mind. Could you share more context in PR description?

Vampire:
Done

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/domain/ActionCoords.kt
```diff
@@ -10,24 +10,26 @@ public data class ActionCoords(
  * A top-level action is an action with its `action.y(a)ml` file in the repository root, as opposed to actions stored
  * in subdirectories.
  */
-public val ActionCoords.isTopLevel: Boolean get() = "/" !in name
+public val ActionCoords.isTopLevel: Boolean get() = "__" !in name
```
krzema12, line 13: I'd like `__` to stay solely as something used when requesting the actions via the Maven-compatible server. It's basically a workaround to avoid a clash with path separators when exposing the action-binding-generator as a Maven server. The `ActionCoords` class along with its extensions belong to the domain layer, so we'd like to stay as close to usual conventions as possible.

I thin the real problem is that it's ambiguous what an action's name is. For non-toplevel actions, is it just the repo name, or any path in the repo as well?

Thinking aloud: maybe instead of keeping the parts after `/` or `__` in the `name`, we should have an extra field right in `ActionCoords` called `path` (right now it's an extension called `ActionCoords.subName`)? By default it would be null, meaning that it's a top-level class, making `ActionCoords.isTopLevel` obvious to implement. Then also e.g. `ActionCoords.repoName` would in fact be equal to `ActionCoords.name`.

Vampire: I'm fine with having always `/` and with having separate additional fields, as long as it is used consistently. :-)

action-binding-generator/src/test/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/generation/GenerationTest.kt
```diff
@@ -287,6 +287,30 @@ class GenerationTest :
             binding.shouldContainAndMatchFile("ActionWithNoInputs.kt")
         }
 
+        test("subaction") {
```
krzema12, line 290: Thanks for adding this missing test 👍

krzema12 (approved):
Thanks for unifying stuff and also simplifying some pieces!

PR #1662, opened 2024-09-24
fix(abg): protect against billion laughs attack

Vampire:


Vampire:
* **#1662** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1662?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypesProviding.kt
```diff
@@ -79,7 +80,7 @@ private fun ActionCoords.fetchTypingsForOlderVersionFromCatalog(fetchUri: (URI)
         } catch (e: IOException) {
             return null
         }
-    val metadata = yaml.decodeFromString<CatalogMetadata>(metadataYml)
```
krzema12, line 82: We also need to use it in other places that load the YAML, so at :
* loading the typings provided by action (`ActionCoords.fetchTypingMetadata`)
* loading the regular action's manifest/metadata (MetadataReading.kt)

Vampire: Why? Aliases are not enabled there.

krzema12: Sorry, right!

action-binding-generator/src/test/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypesProvidingTest.kt
```diff
@@ -508,5 +510,35 @@ class TypesProvidingTest :
                         TypingActualSource.TYPING_CATALOG,
                     )
             }
+
+            test("billion laughs attack is prevented") {
```
krzema12, line 514: If this test fails (i.e. we are again prone to the billion laughs attack by some reason), it hangs and consumes lots of memory. I tried adjusting it like this to make it fail fast:

```suggestion
            test("billion laughs attack is prevented").config(blockingTest = true, timeout = 1.seconds) {
```

but it somehow still hangs. Do you know why? I'm not super-insisting to have it, but I just want to avoid hard-to-debug issues where our CI times out or on local machine, it goes wild 😄

krzema12: @LeoColman do you happen to know the answer or have some hints here?

krzema12: Ok, let's merge as is because it's important to have this fix in place, and maybe follow up on the test's timeout one day.

krzema12 (approved):

action-binding-generator/src/test/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypesProvidingTest.kt
Vampire: If you read the docs, it should become clear.

If you do `blockingTest = true, timeout = 1.seconds` it means the test is executed on a different thread that is then interrupted if the timeout hit.
But there is nothing that would care about that interruption, especially as there is no blocking code executing.

If you do `blockingTest = false, timeout = 1.seconds` it means coroutine means to cancel are used which here also do not apply.

No idea whether there is a way to abort this cleanly using Kotlin.
But at least if you execute the tests with Gradle, you get a pretty clear `OutOfMemoryError` during execution of that test.

krzema12: > But there is nothing that would care about that interruption, especially as there is no blocking code executing.

Why? I understand that the attack creates a very large structure in memory, and creating it takes time and is blocking, it keeps the thread busy. Unless I don't understand the term "blocking" 😊

Vampire: > Unless I don't understand the term "blocking" 

Exactly. :-)

You confuse "blocking" with "long-running".
Blocking is something that does not do work but block.
Something like `Thread.sleep`, or blocking IO operations waiting for bytes coming in over a wire, or waiting for something to connect to a server socket and so on, or waiting to enter some synchronized code part.

The parsing is not blocking, it is just long-running.
It needs significant time, but it continuously does work and does not block.

PR #1831, opened 2025-02-21
fix: fix erroneous version parsing and comparison

Vampire:


Vampire:
* **#1831** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1831?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1831?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Vampire:
Before for example `v2` was parsed as `major`, `minor`, and `patch` all being `2`.
The code was tailored to exclusively being used for three-part versions, not more not less.
But it was also used for major versions for example.

shared-internal/src/test/kotlin/io/github/typesafegithub/workflows/shared/internal/model/VersionTest.kt
```diff
@@ -1,22 +1,42 @@
...
+                ).forEach { (version, isMajor) ->
+                    test("isMajorVersion works correctly for $version") {
+                        Version(version).isMajorVersion() shouldBe isMajor
+                    }
                 }
             }
-        }
-    })
+
+            context("compareTo") {
```
krzema12, line 25: Could we also have a test that shows what inside of `major`, `minor` and `patch` for various cases like below? If we had it earlier, it would show the bug right away. It's true that this test will hook up to some internals that aren't directly used by the clients of this class, but I think it will hell us ensure that parsing works correctly.

krzema12:
Good catch, thanks! Just one request.

shared-internal/src/test/kotlin/io/github/typesafegithub/workflows/shared/internal/model/VersionTest.kt
krzema12: Let's merge this first: https://github.com/typesafegithub/github-workflows-kt/pull/1833

krzema12 (approved):
Thanks!

PR #1835, opened 2025-02-21
feat(server): replace cache library with Aedile

LeoColman:
In order to have a better monitoring of our cache, a library that easily
 exposes metrics is a requirement.

This commit replaces the current cache implementation with Aedile, a wrapper to Caffeine.

Caffeine supports monitoring natively, while aedile allows us to access it without losing the kotlin-like style.


krzema12:
@LeoColman makes sense, thanks! Quick question: did you have a chance to test it, I. e. confirm that metrics appear under the /metrics endpoint? I'm on the go, unable to test.

LeoColman:
image
@krzema12

krzema12 (approved):

PR #1881, opened 2025-04-01
feat(server): log full ktor-client requests on TRACE log level

Vampire:


Vampire:
* **#1881** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1881?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1881?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jit-binding-server/build.gradle.kts
```diff
@@ -10,10 +10,6 @@ dependencies {
     implementation(platform("io.ktor:ktor-bom:3.1.2"))
     implementation("io.ktor:ktor-server-core")
     implementation("io.ktor:ktor-server-netty")
-    implementation("io.ktor:ktor-client-core")
```
krzema12, line 13: These are unused deps right? Thanks for removing them :)

shared-internal/src/main/kotlin/io/github/typesafegithub/workflows/shared/internal/GitHubApp.kt
```diff
@@ -54,6 +60,16 @@ private data class Token(
 
 private val httpClient =
     HttpClient {
+        val klogger = logger
+        install(Logging) {
+            logger =
+                object : Logger {
+                    override fun log(message: String) {
+                        klogger.trace { message }
```
krzema12, line 68: Does it log also headers, including sensitive info like access tokens?

jit-binding-server/build.gradle.kts
Vampire: Yes, ktor-client is only used in the shared-internal code.

shared-internal/src/main/kotlin/io/github/typesafegithub/workflows/shared/internal/GitHubApp.kt
Vampire: Yes, how it is configured it does.
That was intentional, as I also made it only happen on TRACE (by default it logs on INFO what you enable to be logged).
That way you can enable TRACE, see exactly what happens and can also try to do the same request.
If you are concerned, about that, there is also a `sanitizeHeader` option to care about exactly that to clear out certain sensitive headers.

krzema12: Hmm, ok - let's go forward with the current code, since at least right now the tokens are used to just read data, not write anything. It's unlikely to change in the nearest future.

Vampire: Besides that when switching to hub4j ktor-client is gone completely anyway, this is just for the meantime if something happens again, to be able to get a hand on what is happening.

PR #1924, opened 2025-04-29
fix(server): treat failed binding generation as internal server error

Vampire:
Before this PR, it was treated equally if a binding could not be found,
and if some exception happened during initial creating of the binding.
The exceptions were swallowed and the client received a 404 response.
If there was for example a compile error you already got internal server error.
But if for example Kotlin Poet throws an illegal argument exception,
then this did not propagate properly

Now exceptions are logged and the client gets an internal server error
if some exception happened and only a 404 if indeed the action was not found.

One such example is, if a typing has an `enum` with some option that contains a colon.
This resulted in an identifier generated that contains a colon which then produces an
exception due to the illegal identifier character.

### DIFF
--- a/jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/ArtifactRoutes.kt
+++ b/jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/ArtifactRoutes.kt
@@ -31,7 +31,7 @@ import kotlin.time.Duration.Companion.hours
 
 private val logger = logger { }
 
-typealias ArtifactResult = Result<Map<String, Artifact>>
+typealias ArtifactResult = Result<Map<String, Artifact>?>
 
 private val prefetchScope = CoroutineScope(Dispatchers.IO)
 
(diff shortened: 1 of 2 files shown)


Vampire:
* **#1924** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1924?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/ArtifactRoutes.kt
```diff
@@ -40,7 +40,7 @@ private val bindingsCache =
         .newBuilder()
         .refreshAfterWrite(1.hours)
         .recordStats()
-        .asLoadingCache<ActionCoords, ArtifactResult> { runCatching { it.buildVersionArtifacts()!! } }
+        .asLoadingCache<ActionCoords, ArtifactResult> { runCatching { it.buildVersionArtifacts() } }
```
krzema12, line 43: Would be nice to cover it in tests. I believe this part has coverage only in the integration tests, right? Even if it's going to take a bit more time, I'd like us to have it covered in some way, I can help with refactoring if needed.

krzema12:
Good catch! 🚀

jit-binding-server/src/main/kotlin/io/github/typesafegithub/workflows/jitbindingserver/ArtifactRoutes.kt
Vampire: You are welcome to come up with a test. 
I thought about how to test it. 
But to test it, you need to forcibly have a bug in the binding generation, so that an exception happens during generation.

krzema12:
Thanks for this, can't wait to see the failure metric once this fix is deployed 😛

PR #1936, opened 2025-05-01
fix(abg): generate unique enum value names

Vampire:


Vampire:
* **#1936** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1936?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1936?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypingGeneration.kt
```diff
@@ -76,8 +76,12 @@ private fun EnumTyping.buildEnumCustomType(
...
     className: String,
 ): TypeSpec {
-    val itemsNames = itemsNames ?: items.map { it.toPascalCase() }
-    val itemsNameMap = items.zip(itemsNames).toMap()
+    val itemsNames =
+        itemsNames ?: items
+            .map { it.toPascalCase() }
+            .groupBy { it }
+            .values
+            .flatMap { it.mapIndexed { index, name -> "$name${if (index > 0) (index + 1) else ""}" } }
```
krzema12, line 84: How about going with:
```suggestion
            .flatMap { it.mapIndexed { index, name -> "$name_${if (index > 0) (index + 1) else ""}" } }
```
to visually separate the extra part? Especially if the input/output ends with a number (which is rare, but it does in our tests 😄).

krzema12 (changes requested):

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypingGeneration.kt
Vampire: Besides that your change does not compile, I get the point.
I indeed thought about the same and decided against, but it was a close decision without strong opinion, so I changed it to contain the underscore now. :-)
If a value ends in digits like in the test this is most prominent actually.

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypingGeneration.kt
```diff
@@ -76,8 +76,12 @@ private fun EnumTyping.buildEnumCustomType(
     fieldName: String,
     className: String,
 ): TypeSpec {
-    val itemsNames = itemsNames ?: items.map { it.toPascalCase() }
-    val itemsNameMap = items.zip(itemsNames).toMap()
+    val itemsNames =
+        itemsNames ?: items
+            .map { it.toPascalCase() }
+            .groupBy { it }
+            .values
```
krzema12, line 83: One more consideration: since the order of possible enum values doesn't matter, we should be resilient to reordering to not renumber the items. Today, `bam-123` will be accessible under `Bam123_2`, and tomorrow under `Bam123_3`, which may make some workflow either fail or silently succeed in an incorrect way.

How about sorting the values?

krzema12 (changes requested):

action-binding-generator/src/main/kotlin/io/github/typesafegithub/workflows/actionbindinggenerator/typing/TypingGeneration.kt
Vampire: Order in the YAML should be preserved I think, but yeah, why not.
Sorting is added.

krzema12 (approved):

PR #1937, opened 2025-05-01
fix(abg): fix the GMM content to contain sizes, checksums, and the sources variant

Vampire:


Vampire:
* **#1937** <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typesafegithub/github-workflows-kt/1937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/ModuleBuilding.kt
```diff
@@ -2,64 +2,121 @@ package io.github.typesafegithub.workflows.mavenbinding
...
-          "org.gradle.status": "release"
-        }
-      },
-      "createdBy": {
-        "gradle": {
-          "version": "8.7"
-        }
-      },
-      "variants": [
+internal fun ActionCoords.buildModuleFile(
```
krzema12, line 5: Any reason for this change? Is something not working because of this, or is it just to make things right?

Vampire: IntelliJ for example uses the sources variant to get sources and not just tries to get the sources jar, others might do the same.

So my initial intention was to only add the sources variant properly.

While doing so, I recognized that neither the checksums are set, nor the size.
While I did not see anything failing due to this, I strongly assume, that if a consumer - like Gradle - uses the GMM and there is not checksum inside, not checksum check is done and thus download errors or whatever would be caught by the checksums are not checked and thus not caught, so I thought it is better to also fix that right away to deliver correct content.

krzema12: Ok, got it - makes sense! Thanks for being proactive!

maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/VersionArtifactsBuilding.kt
```diff
@@ -12,22 +12,45 @@ data class JarArtifact(
...
+        val mainJarMd5Checksum = { jars.mainJar().md5Checksum() }
+        val mainJarSha1Checksum = { jars.mainJar().sha1Checksum() }
+        val mainJarSha256Checksum = { jars.mainJar().sha256Checksum() }
+        val mainJarSha512Checksum = { jars.mainJar().sha512Checksum() }
+        val sourcesJarSize = { jars.sourcesJar().size }
+        val sourcesJarMd5Checksum = { jars.sourcesJar().md5Checksum() }
+        val sourcesJarSha1Checksum = { jars.sourcesJar().sha1Checksum() }
+        val sourcesJarSha256Checksum = { jars.sourcesJar().sha256Checksum() }
+        val sourcesJarSha512Checksum = { jars.sourcesJar().sha512Checksum() }
+        val module by lazy {
```
krzema12, line 29: I cannot figure out why we need `lazy` here 🤔 How about adding a comment?

maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/ModuleBuilding.kt
```diff
@@ -2,64 +2,121 @@ package io.github.typesafegithub.workflows.mavenbinding
...
+              },
+              "dependencies": [
+                {
+                  "group": "io.github.typesafegithub",
+                  "module": "github-workflows-kt",
+                  "version": {
+                    "requires": "$LATEST_RELASED_LIBRARY_VERSION"
+                  }
+                }
+              ],
```
krzema12, line 86: This is basically for consistency with POM: https://github.com/typesafegithub/github-workflows-kt/blob/76227818c62ce9903ca88116005602a9d999817f/maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/PomBuilding.kt#L26-L31 right?

Vampire: Exactly

maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/VersionArtifactsBuilding.kt
Vampire: To not compile sources and calculate sizes and checksums and so on prematurely.
As soon as `buildModuleFile` is called, the values are needed to build the contents.
With the lazy it is only done when needed.

Alternatively we could also make it a closure and call the closure in the `TextArtifact { ... }`, but this would then also calculate the checksums each time the GMM is requested while with lazy it is only done the first time it is requested.

Actually, now that I think about it, also the lines above should be lazy instead, so that the checksums are only calculated once, even if the GMM and the checksum file requested or requested multiple times.

Currently the checksums are calculated each time they are needed.

krzema12: > Actually, now that I think about it, also the lines above should be lazy instead, so that the checksums are only calculated once, even if the GMM and the checksum file requested or requested multiple times.

True, makes sense!
OK, let's merge as is, and make any improvements as a follow-up.

krzema12 (dismissed):

krzema12:
Hmm, why doesn't CI start?

maven-binding-builder/src/main/kotlin/io/github/typesafegithub/workflows/mavenbinding/VersionArtifactsBuilding.kt
Vampire: Too late, sorry, lazified already incorporated now. :-D

krzema12 (approved):

PR #2298, opened 2026-04-22
fix(server): prefetch the correct binding artifacts

Vampire:

PR #2308, opened 2026-04-25
feat(server): Add commit version significance

Vampire:


Vampire:
* **#2308** <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2308?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2308?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2298** <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2298?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#1602](https://github.com/typesafegithub/github-workflows-kt/pull/1602) <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2309](https://github.com/typesafegithub/github-workflows-kt/pull/2309) <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2309?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

Vampire:
Still missing tests and docs, so still in Draft, but the production changes should be fine already to add the `commit` version significance.  
You specify the normal version in the workflow script, but it writes out the commit id as version and the version as comment.

If no consistency check is done, it just runs with the old commit id.

If a consistency check it done, it fails the consistency check if the version points to another commit id, so this should only be used with concrete non-changeable versios, not with `v1`, but `v1.2.3`.

This also works properly with Renovate, as Renovate will update the commit id in the YAML and the version in the workflow script (if you configured it to do both chagnes in one PR) and it will build successfully, I tried locally.

Vampire:
Ok, I think now it is also ready :-)

.github/workflows/test-script-consuming-jit-bindings.main.do-not-compile.kts
```diff
@@ -24,14 +24,18 @@
 // Always untyped action.
 @file:DependsOn("typesafegithub:always-untyped-action-for-tests:v1")
 
-// Action version pinned to a commit.
+// Action version pinned to an explicit commit.
 @file:DependsOn("actions:setup-python___commit_lenient:v6.0.0__e797f83bcb11b83ae66e0230d6156d7c80228e7c")
 
+// Action version pinned to a commit.
+@file:DependsOn("actions:github-script___commit:v9.0.0")
```
krzema12, line 31: Thanks for the PR! It's a good foundation for what we need, but the design doc for this feature specifies a slightly different API: https://github.com/typesafegithub/design-docs/blob/main/2025-04-27%20Enhancements%20to%20referring%20to%20actions%20by%20commit%20hash.md. So I wouldn't skip specifying commit ID if the user uses `___commit` because the idea is to cross-check the version and the commit ID - they should match. 

However, what you proposed in this PR is also an interesting approach. Maybe we could expose it under `___commit_infer` or sth?

I'd also like to ask for adding consuming an action this way to some tests that surface how the YAML looks like, so likely to `.github/workflows/end-to-end-tests.main.kts`. I know the tests for `___commit_lenient` are missing - my bad, they should be there.

PR #2339, opened 2026-06-07
chore(library): add Detekt rule for forbidding enums

krzema12:
Part of https://github.com/typesafegithub/github-workflows-kt/issues/1578.

The goal is to prevent us from using enums that are problematic when it comes to using values that aren't known when publishing a given version of the library. We should generally use sealed interfaces/classes with sth like `class Custom(val value: String)`. This Detekt rule can be skipped for legitimate cases, on a case basis.

### DIFF
--- a/github-workflows-kt/build.gradle.kts
+++ b/github-workflows-kt/build.gradle.kts
@@ -19,6 +19,11 @@ plugins {
     id("org.jetbrains.dokka") version "2.2.0"
 }
 
+detekt {
+    buildUponDefaultConfig = true
+    config.setFrom("$projectDir/detekt.yml")
+}
+
 group = rootProject.group
 version = rootProject.version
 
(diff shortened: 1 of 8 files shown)


krzema12:
* **#2339** <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2339?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2340** <a href="https://app.graphite.com/github/pr/typesafegithub/github-workflows-kt/2340?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #2353, opened 2026-06-14
feat(library): support types for all triggers that support them

krzema12:
Implements #1784.

Adds a first-class `types` field with a typed sealed `EventType` class to all 18 triggers that support activity types per the [GitHub Actions docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows).

Previously, users had to use `_customArguments = mapOf("types" to ...)` to specify activity types for these triggers.

### DIFF
--- a/github-workflows-kt/src/main/kotlin/io/github/typesafegithub/workflows/yaml/TriggersToYaml.kt
+++ b/github-workflows-kt/src/main/kotlin/io/github/typesafegithub/workflows/yaml/TriggersToYaml.kt
@@ -54,11 +54,29 @@ internal fun List<Trigger>.triggersToYaml(): Map<String, Any?> =
 @InternalGithubActionsApi
 public fun Trigger.toMap(): Map<String, List<String>> =
     when (this) {
-        is Push -> toMap()
+        is BranchProtectionRule -> toMap()
+        is CheckRun -> toMap()
+        is CheckSuite -> toMap()
+        is Discussion -> toMap()
+        is DiscussionComment -> toMap()
+        is IssueComment -> toMap()
+        is Issues -> toMap()
+        is Label -> toMap()
+        is MergeGroup -> toMap()
+        is Milestone -> toMap()
+        is Project -> toMap()
+        is ProjectCard -> toMap()
+        is ProjectColumn -> toMap()
         is PullRequest -> toMap()
+        is PullRequestReview -> toMap()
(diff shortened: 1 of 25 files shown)


github-workflows-kt/src/test/kotlin/io/github/typesafegithub/workflows/domain/triggers/OtherTriggersTest.kt
```diff
@@ -147,7 +184,7 @@ class OtherTriggersTest :
                     "gollum" to emptyMap<Any, Any>(),
                     "issue_comment" to mapOf("types" to listOf("created", "edited", "deleted")),
                     "issues" to mapOf("types" to listOf("opened", "edited")),
-                    "label" to mapOf("types" to listOf("commented", "deleted", "edited")),
```
krzema12, line 150: Why this change? Is it correct?

github-workflows-kt/src/main/kotlin/io/github/typesafegithub/workflows/yaml/TriggersToYaml.kt
```diff
@@ -95,11 +114,101 @@ private fun RepositoryDispatch.toMap() =
         "types" to types,
     )
 
+private fun BranchProtectionRule.toMap() =
```
krzema12, line 117: Maybe we could avoid boilerplate.

github-workflows-kt/src/test/kotlin/io/github/typesafegithub/workflows/domain/triggers/OtherTriggersTest.kt
krzema12: Yes, it's actually fixing a bug we had in tests, see https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#label
But it's not a big deal since this was just a value passed to a custom attribute - the test just showed that it got passed to YAML correctly.

github-workflows-kt/src/main/kotlin/io/github/typesafegithub/workflows/yaml/TriggersToYaml.kt
krzema12: Let's leave it in an explicit form for now.

